PR #23, opened 2020-09-22
Disable elisp-lint by default (fixes #22)

### LINKED ISSUE #22: Disable elisp-lint by default
It almost never works good in a freshly initialized project. It has a bad configuration system and/or default config. Right now it reports 3298 warnings in eldev itself, most of which are caused by misconfiguration, and the rest are also detected by other linters. I think these reasons are enough to disable elisp-lint by default.

sirikid:


eldev.el
```diff
@@ -3050,15 +3050,37 @@ error.")
...
 
 (defmacro eldev-deflinter (name arguments &rest body)
-  "Register a linter in Eldev."
+  "Register a linter in Eldev.
+
+BODY can contain the following keywords:
+
+    :name NAME
+
+        Cleaner name (a symbol) for ther command line.  Default
```
doublep, line 3069: Forgot to change to "Linter name".

eldev.el
```diff
@@ -3327,6 +3356,7 @@ least one warning."
 also runs `package-lint', so you either shouldn't run both or
 change `elisp-lint's configuration."
   :aliases        (elisp-lint el)
+  :default        nil
```
doublep, line 3359: Just remove this line, it's fine to rely on `nil` being the default.

doublep (changes requested):

eldev.el
sirikid: Lol

doublep (approved):

eldev.el
```diff
@@ -3137,8 +3159,12 @@ least one warning."
       (eldev-print "Nothing to do"))))
 
 (defun eldev-lint-default-p (linter)
-  (and (or (eq eldev-lint-default t) (memq linter eldev-lint-default))
-       (not (memq linter eldev-lint-default-excluded)) (not (memq linter eldev-lint-disabled))))
+  (and (eldev-pcase-exhaustive eldev-lint-default
+         (`t t)
+         (:default (assq :default (assq linter eldev--linters)))
```
doublep, line 3164: As it stands, I get the following if I try running it:

    ~/eldev$ eldev lint
    Wrong type argument: listp, (elisp . eldev-linter-elisp)

doublep (changes requested):

sirikid:
I think I fixed it, but I didn't add the test.

doublep (approved):

doublep:
> but I didn't add the test

Doesn't seem to be really needed here, so it's OK.

Seems to work fine now. Also doesn't break Cider.

Merged, thank you.

PR #24, opened 2020-09-23
Fix message for disabled linter

sirikid:
I found this  while tinkering with #23.


doublep:
Fix is fine overall, but I think it's better to reformulate the test without creating yet another project. You can use option `--setup` to add one-off configuration, so that you don't need file `Eldev` every time you need something slightly different. E.g. see test `eldev-command-hook-1`.

doublep:
Please squash the commits and I will merge this. In general, I and some/many other projects follow the policy that most PR (unless really large and with broad scope) should contain only one commit and thus history can be freely rewritten until they are merged. But if something needs to be fixed after a PR is already accepted, then of course this goes as a separate commit.

BTW, I usually use more lax regexps in testing to avoid insisting on specific wording in messages. But it's fine either way.

doublep:
Thank you.

PR #37, opened 2021-04-21
Add Windows specific documentation

juergenhoetzel:
Feel free to edit. I'm new to `asciidoc` :fearful:

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -167,11 +167,14 @@ There are several ways to install Eldev.
 
 [discrete]
 ==== Bootstrapping from MELPA: if you have a catch-all directory for executables
-
-. From this directory (e.g. `~/bin`) execute:
-+
 --
+* On Unix, from this directory (e.g. `~/bin`) execute:
++
     $ curl -fsSL https://raw.github.com/doublep/eldev/master/bin/eldev > eldev && chmod a+x eldev
+
+* On Windows, from this directory (e.g. `%USERPROFILE%\bin`) execute:
++
+    $ curl.exe -fsSL https://raw.github.com/doublep/eldev/master/bin/eldev.bat > eldev.bat
 
 You can even do this from `/usr/local/bin` provided you have the
 necessary permissions.
@@ -182,6 +185,8 @@ first invocation.
 
 [discrete]
 ==== Bootstrapping from MELPA: general case
+
+===== Unix
 
 . Execute:
 +
@@ -194,6 +199,21 @@ This will install `eldev` script to `~/.eldev/bin`.
 . Add the directory to your `$PATH`; e.g. in `~/.profile` add this:
 +
       export PATH="$HOME/.eldev/bin:$PATH"
+
+===== Windows
+
+. Execute:
++
+--
+    $ curl.exe -fsSL https://raw.github.com/doublep/eldev/master/webinstall/eldev.bat | cmd
+
+This will install `eldev.bat` script to `%USERPROFILE%\.eldev\bin`.
+--
+
+. Add the directory to your `%PATH%`:
++
+      reg add HKCU\Environment /v Path /d "%USERPROFILE%\.eldev\bin;%PATH%" /f
+
 
 Afterwards Eldev will bootstrap itself as needed on first invocation.
 


doublep:
Yeah, I will restructure this somewhat. What is most important are the exact commands.

Some questions:

* I get from the PR that system variables on Windows are `%VAR%` instead of `$VAR`, right?
* Can you create sth. like `install.bat` similar to `install.sh` for the third way? It's not very important though.
* Does `ELDEV_LOCAL` work on Windows? Can it be specified on the command line, like in Linux, e.g. `ELDEV_LOCAL=. bin/eldev ...`? And what effect does your answer have on the fourth way to install?
* Is there any standard prefix for command lines? E.g. using `$` is widely recognized for UNIX-like OSes like this: `$ echo foo`, because non-root shell promt usually ends with `$`.
* Can you provide some commands for CI? Can anything already suggested (EVM, nix-emacs-ci) be used, or does it have to be something completely different?

juergenhoetzel:
>     * I get from the PR that system variables on Windows are `%VAR%` instead of `$VAR`, right?

Exactly
 
>     * Can you create sth. like `install.bat` similar to `install.sh` for the third way? It's not very important though.

@ikappaki  already ported this:
https://github.com/doublep/eldev/blob/master/install.bat

i just forgot to mention it in this PR.
 
>     * Does `ELDEV_LOCAL` work on Windows? Can it be specified on the command line, like in Linux, e.g. `ELDEV_LOCAL=. bin/eldev ...`? And what effect does your answer have on the fourth way to install?
Exactly: 
```
set ELDEV_LOCAL=.
```
also works fine for local development on Windows.

> 
>     * Is there any standard prefix for command lines? E.g. using `$` is widely recognized for UNIX-like OSes like this: `$ echo foo`, because non-root shell promt usually ends with `$`.

This is the default prompt on a Windows 10 System:

> 
>     * Can you provide some commands for CI? Can anything already suggested (EVM, nix-emacs-ci) be used, or does it have to be something completely different?

doublep:
> @ikappaki already ported this

Indeed, so it's all right then. I mistakenly checked in the `future-doc` branch, which doesn't have it yet of course.

> The mentioned solutions don't work on Windows. Someone could write a portable solution (which uses nix on Unix and scoop on Windows).

Is it possible to document what you are doing in Eldev's own CI?

Also forgot to ask. Is it possible to run a command that is not in the current directory and not in `%PATH%` on Windows? E.g. on Linux you could run it `$ path/to/executable ...`. You actually always need a path if the directory is not in `$PATH`. E.g. just `./`, but not an empty string.

doublep:
One more thing. You are writing `curl.exe` etc., but doesn't it work without the extension too on Windows?

juergenhoetzel:
> One more thing. You are writing `curl.exe` etc., but doesn't it work without the extension too on Windows?

Yes, I did that on purpose. Because `curl` is unfortunately also a Powershell command under Windows, which is incompatible with `curl`:
```
PS C:\Users\pidsleywin> Get-Command curl.exe

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     curl.exe                                           7.55.1.0   C:\Windows\system32\curl.exe


PS C:\Users\pidsleywin> Get-Command curl

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Alias           curl -> Invoke-WebRequest
```

doublep:
> Yes, I did that on purpose. Because curl is unfortunately also a Powershell command under Windows, which is incompatible with curl

Understood. Please reply to my previous comment too.

juergenhoetzel:
> > The mentioned solutions don't work on Windows. Someone could write a portable solution (which uses nix on Unix and scoop on Windows).
> 
> Is it possible to document what you are doing in Eldev's own CI?

You mean the javascript code? It is  just used to avoid non-portable shell code. It is actually a 1:1 implementation of the previous `bash` code.

For normal Emacs developers using `eldev` in their CI I propose to use a Github-Action to setup  `eldev` in a declarative way.

I meanwhile wrote a simple GA for this use case:

https://github.com/marketplace/actions/setup-eldev 

It should also work out of the box on Windows when a new release of `Eldev` is pushed to `MELPA stable`. Currently the Windows-jobs are failing in my sample project:

https://github.com/juergenhoetzel/github-eldev-test/actions/runs/782680025 

> Also forgot to ask. Is it possible to run a command that is not in the current directory and not in `%PATH%` on Windows? E.g. on Linux you could run it `$ path/to/executable ...`. You actually always need a path if the directory is not in `$PATH`. E.g. just `./`, but not an empty string.

Yes this works both with Powershell and CMD:

```
PS C:\Users\pidsleywin> C:/Users/pidsleywin/.eldev/bin/eldev version
eldev 0.8.1
```

doublep:
> You mean the javascript code?

No, rather things about Scoop.

> For normal Emacs developers using eldev in their CI I propose to use a Github-Action to setup eldev in a declarative way.

It's a good idea to provide non-GitHub way too, not everyone is using GitHub. I guess Windows installation commands as documented should be usable, but I don't know. E.g. is `curl.exe` available by default or do you need to install it first (via Scoop?) and so on. So I'd rather wait for some comments from someone who uses Windows, not write non-working crap myself.

> I meanwhile wrote a simple GA for this use case:

OK, let's wait for this. More documentation can also be added post-release (`future-doc` is only about unreleased features).

Meanwhile, I committed documentation based on your changes (esp. commands), but different enough that it was easier not to bother with merging etc. See commit a04f52f if you want to check it.

PR #38, opened 2021-04-22
Check if %1 got expanded

juergenhoetzel:
This prevents an invalid URL from being generated when the
webinstaller is used via a cmd.exe pipe.

Fixes this issue:
```
curl: (6) Could not resolve host: %1
```
In the CI tests we don't pipe the `webinstaller\eldev.bat` into `cmd.exe`. 
That's why this error didn't show up in the CI tests.


doublep:
Thank you.

PR #40, opened 2021-04-26
Use jcs090218/setup-emacs-windows

juergenhoetzel:
Use Github Action instead of manually setting up emacs by using powershell/scoop
commands.


doublep:
Merged, thank you.

Issue #43, opened 2021-05-14
Symbol’s function definition is void: uninstallable


DarwinAwardWinner:
I just upgraded to Eldev 0.9, and now most commands fail with the following error:

```
$ eldev version
eldev 0.9
$ eldev -d upgrade # Also happens for compile, test, ...
Debugger entered--Lisp error: (void-function uninstallable)
  uninstallable(nil nil)
  eldev--install-or-upgrade-dependencies(project (test) t nil nil t)
  eldev-upgrade()
  apply(eldev-upgrade nil)
  eldev-cli(("-d" "upgrade"))
  (kill-emacs (eldev-cli (append (cdr (member "--" command-line-args)) nil)))
  eval((kill-emacs (eldev-cli (append (cdr (member "--" command-line-args)) nil))) t)
  command-line-1(("--execute" "(let ((eldev--emacs-version (format \"%s.%s\" emacs-..." "--execute" "(kill-emacs (eldev-cli (append (cdr (member \"--\" c..." "--" "-d" "upgrade"))
  command-line()
  normal-top-level()
```

Looking at the source, it seems this value is bound using a special eldev lazy evaluation macro, so I'm not sure how to debug it.

DarwinAwardWinner:
Deleting `~/.eldev` and re-running seems to have fixed this. I guess it's another ghost of upgrades past.

doublep:
Good that it got resolved. Do you remember which version you had had before upgrading? Maybe it is reproducible and should be properly fixed in that case.

DarwinAwardWinner:
I don't remember exactly, but it started with 0.8. FWIW I just tried on another computer upgrading from 0.8.1 to 0.9, and the issue didn't happen.

doublep:
Would be nice to have a copy of `~/.eldev` contents, or at least the list of contained files, but I guess it's too late now. I'll keep this issue open in case it happens again.

DarwinAwardWinner:
Hmm, actually, I have Time Machine enabled, let me see if I can pull the previous contents from a backup.

DarwinAwardWinner:
Ok, I recovered the offending `~/.eldev` contents, and I can reproduce the error:

```
$ emacs --version
GNU Emacs 27.2
Copyright (C) 2021 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of GNU Emacs
under the terms of the GNU General Public License.
[...]

PR #46, opened 2021-06-19
Inhibit end-of-line conversion in tar-mode buffer

juergenhoetzel:
This is a workaround to prevent corrupt tar files on windows.

This reverts commit 9e59d6 (Disable the tests added in commit 1f093ac on Windows, because of an issue unrelated to Eldev)

Refs #43


doublep:
Thank you, merged.

PR #47, opened 2021-07-01
Fix argument parsing issue for Windows github web-installer

juergenhoetzel:
Command line arguments %[1-9] do not exist when piping to CMD.

This is a follow up for https://github.com/doublep/eldev/pull/41 where we missed to fix the issue also for the github webinstaller.

### DIFF
--- a/webinstall/github-eldev.bat
+++ b/webinstall/github-eldev.bat
@@ -1,24 +1,27 @@
-rem This script downloads Eldev startup script as `USERPROFILE/.eldev/bin/eldev'
-rem for use in a GitHub workflow.
+@echo off
+rem This script downloads Eldev startup script as `%USERPROFILE%/.eldev/bin/eldev'.
 rem
 rem curl.exe is included in Windows 10 since April-2018-Update (1803)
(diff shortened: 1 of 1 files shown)


doublep:
It looks a bit strange that the file is completely replaced according to the diff. Was it really meant to be like that?

doublep:
The failure is not in your patch, it's already in `master`. Not sure if you received notification from commit 62648d3. But this is on Windows only either way, so would be nice if you looked into it.

juergenhoetzel:
> The failure is not in your patch, it's already in `master`. Not sure if you received notification from commit [62648d3](https://github.com/doublep/eldev/commit/62648d350439cffa3c87461bbaf0414f87136d35). But this is on Windows only either way, so would be nice if you looked into it.

I will take a look.

juergenhoetzel:
> It looks a bit strange that the file is completely replaced according to the diff. Was it really meant to be like that?

Yes I have  consolidated the two webinstall scripts (the github version did also still use CRLF line endings). I should have mentioned this in the commit message: I will amend the commit message.

doublep:
Thank you, merging.

PR #48, opened 2021-07-09
Fix windows script update

juergenhoetzel:
This is really an obscure issue. Fixes:

```
             Cannot locate installed script `eldev.ps1', skipping
             Cannot locate installed script `eldev.bat', skipping
```

and 

```
[00:14.519]  Stdout contents:
             Upgraded or installed 1 package
             Upgraded script `c:/Users/PIDSLE~1/AppData/Local/Temp/eldev-xoazUY-test-bin/bin/eldev.ps1'
             Upgraded script `c:/Users/PIDSLE~1/AppData/Local/Temp/eldev-xoazUY-test-bin/bin/eldev.bat'

[00:14.530]  Stderr contents:
             [1/1] Upgrading package `eldev' (0.9.2snapshot -> 999.9) from `bootstrap-pa'...
             Der Befehl "--" ist entweder falsch geschrieben oder
             konnte nicht gefunden werden.

```
Appending the "rem" comment instead of prepending fixes the issue.

Running a `.bat` script  while it is updating itself seems to lead to obscure errors as listed above.

### DIFF
--- a/eldev.el
+++ b/eldev.el
@@ -2237,7 +2237,7 @@ Since 0.2."
     (dolist (script scripts)
       (catch 'continue
         (let* ((installed      (eldev-shell-command))
-               (installed-name (file-name-base installed)))
+               (installed-name (file-name-nondirectory installed)))
           (unless (string= installed-name script)
             (if (member installed-name scripts)
                 (setf installed (expand-file-name script (file-name-directory installed)))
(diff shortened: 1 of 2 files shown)


doublep:
Thank you! Using `file-name-base` instead of `file-name-nondirectory` was really my mistake, I just never noticed since on non-Windows the file is called without extension, so they both give the same result in that specific case. But some other things I wouldn't be able to catch without access to Windows at all.

PR #53, opened 2021-10-20
Add Command for Running Emacs in Docker

LaurenceWarne:
Hi!  I've been playing around a bit for a way to run my projects with different Emacs versions without having to have multiple Emacs installations and/or something like evm.

I thought it might be useful to take advantage of the existing eldev docker images on https://hub.docker.com/r/silex/emacs to do this.  Demo:

https://user-images.githubusercontent.com/17688577/138159777-9cd027a6-5753-4b2d-b3b4-c6f333ed7e8d.mp4

Essentially the command just runs an appropriate docker image and then executes `eldev emacs` with the rest of the passed args.  Would you be interested in this?  - No problem if you think this would be more appropriate as a plugin or something.

I've left out automated testing and documentation, happy to add them (to this PR) if you're interested in this change.

### DIFF
--- a/eldev-util.el
+++ b/eldev-util.el
@@ -990,6 +990,10 @@ through `executable-find' if possible.  Since Eldev 0.8.")
 Can be set explicitly or left to t, in which case it is located
 through `executable-find' if possible.  Since Eldev 0.8.")
 
+(defvar eldev-docker-executable t
+  "Docker executable.
+Can be set explicitly or left to t, in which case it is located
+through `executable-find' if possible.")
 
 (defmacro eldev-find-executable (cache-var not-required finder-form error-message &rest error-arguments)
   "Find and executable using FINDER-FORM.
@@ -1071,6 +1075,13 @@ See also variable `eldev-svnadmin-executable'."
   (eldev-find-executable eldev-svnadmin-executable not-required
     (executable-find "svnadmin")
     "Subversion is not installed (cannot find `svnadmin' executable)"))
+
+(defun eldev-docker-executable (&optional not-required)
+  "Find `docker' executable.
+See also variable `eldev-docker-executable'."
+  (eldev-find-executable eldev-docker-executable not-required
+    (executable-find "docker")
+    "Docker is not installed (cannot find `docker' executable)"))
 
 (defvar vc-svn-program)
 (with-eval-after-load 'vc-svn
@@ -1164,7 +1175,7 @@ Also, eat up several options from BODY if present:
       (eldev-verbose header-message))
     (if only-when-verbose
         (eldev-verbose "%s" (buffer-string))
-      (eldev-output "%s" (buffer-string)))))
+      (eldev-output :nolf "%s" (buffer-string)))))
 
 
 
--- a/eldev.el
+++ b/eldev.el
@@ -111,6 +111,9 @@ instead.")
   "Name of Eldev cache subdirectory, `.eldev'.
 See also function `eldev-cache-dir'.")
 
+(defconst eldev-global-cache-dir "global-cache"
+  "Name of the global cache directory (a subdirectory of `eldev-dir').")
+
 (defvar eldev--internal-pseudoarchive "--eldev--")
 
 (defvar eldev--loading-modes
@@ -980,7 +983,7 @@ Returns COMMAND-LINE with options removed."
       (nreverse without-options))))
 
 (defun eldev-global-package-archive-cache-dir ()
-  (expand-file-name "global-cache" eldev-dir))
+  (expand-file-name eldev-global-cache-dir eldev-dir))
 
 (defun eldev-cache-dir (emacs-version-specific &optional ensure-exists)
   "Get the directory where various internal caches should be stored.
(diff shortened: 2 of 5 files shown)


doublep:
> I created antoher image: [LaurenceWarne@aadfcd7](https://github.com/LaurenceWarne/eldev/commit/aadfcd7f3c9b075e6e965821e77130c81ded39dd), it works by creating another user with the same uid on the container as the host, and running with that user on the container

Uh, is there no easier way? I also searched for this on the Internet briefly, and haven't found anything simple. But is it really not possible to go without a specialized image? This would effectively remove the option to use user's own image (i.e. what we discussed before, either VER -> silex's image, or IMAGE-NAME, for a custom one), or at least require that they base on "our" image somehow...

> It lgtm testing, in particular it solves the permissions errors on the test runs. lmk what you think of this - I'm not sure where it would live (I don't think it should be part of the default `ci-eldev` img since it does a lot of stuff CI users probably shouldn't care about).

Yeah, I thought about negotiating with silex to include this into his images, but now I see that there are quite a few things going in it.

eldev.el
```diff
@@ -4028,6 +4028,90 @@ be passed to Emacs, else it will most likely fail."
...
+  (let ((img (funcall eldev-docker-img-fn (car parameters))))
+    (eldev-call-process
+     eldev-docker-executable
+     (list "pull" img)
+     :pre-execution (eldev-verbose "Pulling image %s" img)
+     :die-on-error (format "%s pull" eldev-docker-executable)
+     (eldev--forward-process-output
+      (format "Output of %s pull:" eldev-docker-executable)
+      (format "%s process produced no output" eldev-docker-executable))
+    (let* ((as-gui (not (member "--batch" parameters)))
```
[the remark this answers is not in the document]
doublep: I meant, what about using a different command in the Docker, not `emacs`? E.g. this unrealistic, but working example: `eldev eval -- "(setf --batch t)" --batch` -> `eldev docker 26 eval -- "(setf --batch t)" --batch`. In principle, this does include `--batch` on the command line...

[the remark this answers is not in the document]
LaurenceWarne: Sorry I'm not sure I follow, both the above commands give the same output for me (minus the bootstrapping output for docker)?

I was under the impression that Emacs should only be started as a GUI (equivalently, we add `-e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix silex/emacs` to `docker run`) iff the eldev command is `emacs` and `--batch` is not passed, is this correct?

LaurenceWarne:
> Uh, is there no easier way? I also searched for this on the Internet briefly, and haven't found anything simple. But is it really not possible to go without a specialized image? This would effectively remove the option to use user's own image (i.e. what we discussed before, either VER -> silex's image, or IMAGE-NAME, for a custom one), or at least require that they base on "our" image somehow...

I've just thought of a horrific hack to get away with using the same image (or any image with an Emacs installation): the problem with using `--user UID:GID` is that this 'user' on the container won't have permission to run eldev (or really do anything outside of the mounted volumes), but if we pass a `HOME` variable to some directory on a docker volume, then we can install eldev when `docker run` is called, example (setting the HOME to inside the project directory):
```
docker run -e HOME=/eldev/fake-home -it --rm -u 1000:1000 -v /home/laurencewarne/projects/eldev/:/eldev -w /eldev -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix silex/emacs:27-ci-eldev sh -c 'curl -fsSL https://raw.github.com/doublep/eldev/master/webinstall/eldev | sh && export PATH="/eldev/fake-home/.eldev/bin:$PATH" && eldev test'
```
This appear s to work, though seems a bit fragile, and we'd have to mount the global cache and config inside this new home, etc.  wdyt?

> It would be interesting to test something like eldev docker 26 docker 25 test, i.e. a Docker within a Docker. Largely for fun, but also to potentially spot some otherwise hard-to-discover errors. Since this is going to be slow, really limit this to one test.

We'd have to install docker on the images also :smile:

eldev.el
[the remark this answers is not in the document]
doublep: You either fixed/improved it independetly, or just forgot. I see the current code already does what I was trying to say. So, this is resolved.

doublep:
> I've just thought of a horrific hack

If you have looked at Eldev code enough, you must know that I'm not against hacks. The source code is full of hacks and workarounds. Whatever works, unless it really works "by side effect only" or is likely to break with any small future change in the program we are integrating with here (i.e. Docker).

> [...]
> This appear s to work, though seems a bit fragile, and we'd have to mount the global cache and config inside this new home, etc. I don't know if I'd call this an "easier way", but wdyt?

As it is looks fine to me. I'm only not sure I understand the meaning of "mount the global cache and config inside this new home, etc.".

This fake root should be placed somewhere inside `(eldev-cache-dir t t)`. The binary we can just copy from `(locate-file (format "bin/eldev" script) load-path)`, see `eldev-upgrade-self`.

> I don't know if I'd call this an "easier way"

It is easier if we don't have to create and maintain (remember about rebuilding when a new Emacs version comes) extra images. And even better if it allows users to use their own images. I don't care if it results in 10, 30 or 100 more lines of Elisp code.

> > It would be interesting to test something like eldev docker 26 docker 25 test, i.e. a Docker within a Docker. Largely for fun, but also to potentially spot some otherwise hard-to-discover errors. Since this is going to be slow, really limit this to one test.
> 
> We'd have to install docker on the images also

Crap, didn't think about that. Oh well ;)

LaurenceWarne:
Cool, it looks to have fixed the permissions on CI.

> As it is looks fine to me. I'm only not sure I understand the meaning of "mount the global cache and config inside this new home, etc.".

Just that we have to change the target mount points on the container, e.g. `/root/.eldev/config` -> `/project-dir/.eldev/docker-home/.eldev/config`.

As suggested, the "fake home" on the host is created in `eldev-cache-dir`/`docker-home`.  This directory needs to be created on the host prior to any `eldev docker` call else when docker creates it on the container it is owned by root (see `eldev--docker-create-directories`).  It's currently persisted between runs (so you would be able to see it in `.eldev`).  Do you think this should be deleted post any `eldev docker` run, or do you think something like changing the dir to `eldev-cache-dir`/`.docker-home` would be sufficient?

**TODO**
- Don't install eldev from melpa for the tests, use the system eldev
- General windows compat

doublep:
> Do you think this should be deleted post any eldev docker run, or do you think something like changing the dir to eldev-cache-dir/.docker-home would be sufficient?

It can stay if it contains anything useful that could speed up future Docker runs. Otherwise delete it after all runs. However, future runs must not fail if the directory didn't get deleted for whatever reason. Also, no need to rename it with a dot, because cache directory is already hidden (it is that `.eldev` you have in projects).

> Also we don't use eldev on the eldev-ci images so we can use plain old Emacs images.

Agreed. Also wanted to mention, but I'm too late.

> Don't install eldev from melpa for the tests, use the system eldev

This is very important. Basically, when testing, I want to use Eldev that I'm developing right now for all tests. If I break `eldev-exec` in any way, I want both `eldev exec` and `eldev docker ... exec` blow up immediately.

> General windows compat

This we can forget because of the hack. I guess it would be even better to add an explicit test for `system-type` in the new command and outright fail on systems that don't have user IDs, or at least not in the way we need/expect. I.e. let's only make it work where we are sure it will work. Others can try and make it support Windows later.

doublep:
Looks like CI is broken again, this time on macOS. Let's wait for a while.

LaurenceWarne:
> It can stay if it contains anything useful that could speed up future Docker runs. Otherwise delete it after all runs. However, future runs must not fail if the directory didn't get deleted for whatever reason. Also, no need to rename it with a dot, because cache directory is already hidden (it is that .eldev you have in projects).

I've set it up so it's deleted after runs.  But there should be no problem if it happens to exist prior to a run.

> This is very important. Basically, when testing, I want to use Eldev that I'm developing right now for all tests. If I break eldev-exec in any way, I want both eldev exec and eldev docker ... exec blow up immediately.

Cool the last few commits should handle this.

> This we can forget because of the hack. I guess it would be even better to add an explicit test for system-type in the new command and outright fail on systems that don't have user IDs, or at least not in the way we need/expect. I.e. let's only make it work where we are sure it will work. Others can try and make it support Windows later.

If you're ok with that, I've disabled the tests for windows. I've just noticed that on GH actions, mac os appears not to have docker installed :disappointed:

LaurenceWarne:
> It can stay if it contains anything useful that could speed up future Docker runs. Otherwise delete it after all runs. However, future runs must not fail if the directory didn't get deleted for whatever reason. Also, no need to rename it with a dot, because cache directory is already hidden (it is that .eldev you have in projects).

I've set it up so it's deleted after runs.  But there should be no problem if it happens to exist prior to a run.

> This is very important. Basically, when testing, I want to use Eldev that I'm developing right now for all tests. If I break eldev-exec in any way, I want both eldev exec and eldev docker ... exec blow up immediately.

Cool the last few commits should handle this.

> This we can forget because of the hack. I guess it would be even better to add an explicit test for system-type in the new command and outright fail on systems that don't have user IDs, or at least not in the way we need/expect. I.e. let's only make it work where we are sure it will work. Others can try and make it support Windows later.

If you're ok with that, I've disabled the tests for windows. I've just noticed that on GH actions, mac os appears not to have docker installed :disappointed:

eldev-util.el
```diff
@@ -1091,16 +1102,17 @@ Also, eat up several options from BODY if present:
                      (signal 'eldev-error (list "%s exited with error code %d" (eldev-message-upcase-first description) exit-code))))))
            ,@(or body '(exit-code)))))))
 
-(defun eldev--forward-process-output (&optional header-message header-if-empty-output only-when-verbose)
+(defun eldev--forward-process-output (&optional header-message header-if-empty-output only-when-verbose nolf)
```
doublep, line 1105: Actually, it should always work as if `nolf` was specified, even for existing invocations. Drop the parameter, just change the function. For `eldev-verbose` it should stay as it is, just in case the invoked process doesn't print ending newline. But for `eldev-output` we should just pass generated output without changes, not even supplying trailing `lf`.

eldev.el
```diff
@@ -111,6 +111,9 @@ instead.")
   "Name of Eldev cache subdirectory, `.eldev'.
 See also function `eldev-cache-dir'.")
 
+(defconst eldev-global-cache-directory-name "global-cache"
```
doublep, line 114: Just use `-dir` instead of `-directory-name`, for consistency with e.g. `eldev-cache-dir`. I don't like to abbreviate too much, but sometimes names simply become too long otherwise.

eldev.el
```diff
@@ -4028,6 +4031,221 @@ be passed to Emacs, else it will most likely fail."
...
+It should take one parameter: the arguments of the \"eldev\" call.")
+
+(defvar eldev--docker-home-name "docker-home"
+  "Name of the home directory of the docker user.")
+
+(defvar eldev--docker-os-error-fmt-string
+  "OS %s is not currently supported by \"eldev docker\""
+  "Error message format string if the os is not supported.")
+
+(defun eldev--container-bootstrap-cmd-fn (eldev-args)
```
doublep, line 4060: Can we replace bootstrapping with copying the script file (it is now always bundled with Eldev, so simple `locate-file` call should do)? While this sort of implies that we depend on extact target's OS, it's not like bootstrapping commands are cross-platform either. As an added benefit, we'd have "test the source code, not something published" here too, even if for bootstrap scripts it rarely matters.

eldev.el
```diff
@@ -4028,6 +4031,221 @@ be passed to Emacs, else it will most likely fail."
...
+  (let* ((img (eldev--docker-determine-img (car parameters)))
+         (docker-exec (eldev-docker-executable))
+         (escaped-params (mapcar #'eldev-quote-sh-string (cdr parameters)))
+         (container-cmd (eldev--docker-container-eldev-cmd escaped-params))
+         (as-gui (and (string= "emacs" container-cmd)
+                      (not (member "--batch" parameters))))
+         (eldev--container-bootstrap-cmd-fn
+          (if (getenv "ELDEV_LOCAL")
+              (apply-partially
+               #'eldev--container-eldev-source-install-cmd "/eldev")
```
doublep, line 4226: Is this correct? I don't see the result of `(getenv "ELDEV_LOCAL")` being used anywhere. Maybe I just don't follow the logic.

doublep:
> If you're ok with that, I've disabled the tests for windows. Though I've just noticed that on GH actions, mac os appears not to have docker installed

Yes, fine. Also fine to skip tests when Docker is not found, and as I understand it, your code does just that already. Please rebase on `master`, I skip tests on macOS now, because `setup-emacs` action no longer works there (already reported).

eldev-util.el
LaurenceWarne: So just to double check, the function should look like this? :
```elisp
(defun eldev--forward-process-output (&optional header-message header-if-empty-output only-when-verbose)
  (if (= (point-min) (point-max))
      (when header-if-empty-output
        (eldev-verbose header-if-empty-output))
    (when header-message
      (eldev-verbose header-message))
    (if only-when-verbose
        (eldev-verbose "%s" (buffer-string))
      (eldev-output :nolf "%s" (buffer-string)))))
```
In which case, am I ok to fix the existing tests which expect a newline from this fn, e.g. `eldev-emacs-2`?

doublep: Yes. Adjust all tests that depend on this accordingly, they are not meant to be set in stone.

doublep:
Please address the latest review (four points) and we can finally merge this. Before merging, please squash everything into one commit. Or, if you want, break out stuff that is not related to Docker per se, e.g. there could be a separate commit that adds `:nolf` in `eldev--forward-process-output` and adjusts the tests "broken" because of that.

Also, please remove changes to `README.adoc`. I commit documentation changes to a separate branch, so that what can be seen on the [project page](https://github.com/doublep/eldev) always reflects current stable release, and not ongoing development. If you want, file a separate PR against `future-doc` branch, or else I will just add the documentation manually later.

doublep:
Ah, and also please rebase on `master`, so that I can run tests. Otherwise they would fail because macOS testing is broken for unrelated reasons (in `master` I already disabled it).

eldev.el
LaurenceWarne: The function `eldev--conatainer-eldev-source-install-cmd` is only concerned with the location of the eldev source on the container and assumes that it's mounted in `/eldev`.  In this commit, I was using `(getenv "ELDEV_LOCAL")` as an indicator and calling it again when the mounts are determined, but this is lazy, the most recent changes have it so that `(getenv "ELDEV_LOCAL")` is called only once and its value passed through.

eldev.el
LaurenceWarne: :+1: should be done in the new commit.

test/integration/docker.el
```diff
@@ -0,0 +1,47 @@
...
+         "docker"
+         eldev--docker-emacs-version
+         "emacs"
+         "--batch"
+         "--eval"
+         `(prin1 (+ 1 2)))
+        (should (string-suffix-p "3" stdout))
+        (should (= exit-code 0)))))
+
+(ert-deftest eldev-docker-emacs-2 ()
```
LaurenceWarne, line 20: I've added a new test against Emacs `24`, I thought it might be useful to test against the oldest supported version.  Let me know if you think this is overkill and I'll remove it.

test/integration/docker.el
```diff
@@ -0,0 +1,47 @@
...
+  (let ((eldev--test-eldev-local (expand-file-name default-directory)))
+    (eldev--test-run "trivial-project"
+        ("--quiet"
+         "docker"
+         eldev--docker-emacs-version
+         "emacs"
+         "--batch"
+         "--eval"
+         `(prin1 (+ 1 2)))
+        (should (string-suffix-p "3" stdout))
```
LaurenceWarne, line 17: The stdout is exactly the same as the command w/o docker if the user has the image before hand, but on CI the "not found locally, pulling bla..." will be on the start of stdout.  If you don't like this I can see what can be done, maybe filtering it out in the forwarded output or something.

LaurenceWarne:
> Please address the latest review (four points) and we can finally merge this. Before merging, please squash everything into one commit. Or, if you want, break out stuff that is not related to Docker per se, e.g. there could be a separate commit that adds :nolf in eldev--forward-process-output and adjusts the tests "broken" because of that.

Done!  I've been lazy and stuck everything all in one :slightly_smiling_face: Thanks for your patience.

> Also, please remove changes to README.adoc. I commit documentation changes to a separate branch, so that what can be seen on the project page always reflects current stable release, and not ongoing development. If you want, file a separate PR against future-doc branch, or else I will just add the documentation manually later.

No problem, I can PR with the doc changes I have locally if you like.

doublep:
Hm, the automated tests on GitHub servers pass, but I have problems running anything locally. Neither from Eldev project itself:

    ~/eldev$ eldev -dt docker 27 eval 1
    Started up on Sat Nov 6 19:05:43 2021
    Running on GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 2.24.32)
     of 2020-10-27
    Project directory: ‘/home/paul/eldev/’
    Loading file ‘/home/paul/.eldev/config’...
    Loading file ‘Eldev’...
    Loading file ‘Eldev-local’...
    Executing command ‘docker’...
    Executing ‘eldev-executing-command-hook’...
    Running command ’/usr/bin/docker run --rm -e HOME=/eldev/.eldev/docker-home -u 1000:1000 -v /home/paul/eldev/:/eldev -w /eldev -v ~/eldev:/eldev -v /home/paul/.eldev/config:/eldev/.eldev/docker-home/.eldev/config -v /home/paul/.eldev/global-cache:/eldev/.eldev/docker-home/.eldev/global-cache silex/emacs:27 sh -c ELDEV_LOCAL=/eldev /eldev/bin/eldev eval 1’

    Output of the /usr/bin/docker run:
    docker: Error response from daemon: Duplicate mount point: /eldev.
    See 'docker run --help'.

    /usr/bin/docker run exited with error code 125
    Finished erroneously on Sat Nov 6 19:05:43 2021

nor from another, with a different error:

    ~/iter2$ eldev -dt docker 27 eval 1
    Started up on Sat Nov 6 19:05:38 2021
    Running on GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 2.24.32)
     of 2020-10-27
    Project directory: ‘/home/paul/iter2/’
    Loading file ‘/home/paul/.eldev/config’...
    Loading file ‘Eldev’...
    Loading file ‘Eldev-local’...
    Reading internal evaluation cache from file ‘.eldev/27.1/internal-eval.cache’...
    Using cached value for form ‘eldev-project-main-file’ in directory ‘/home/paul/git/undercover/’: nil
    Will use directory ‘~/git/undercover/’ as local dependency ‘undercover’ with loading mode ‘as-is’
    Executing command ‘docker’...
    Running command ’/usr/bin/docker run --rm -e HOME=/iter2/.eldev/docker-home -u 1000:1000 -v /home/paul/iter2/:/iter2 -w /iter2 -v ~/eldev:/eldev -v /home/paul/.eldev/config:/iter2/.eldev/docker-home/.eldev/config -v /home/paul/.eldev/global-cache:/iter2/.eldev/docker-home/.eldev/global-cache -v /home/paul/git/undercover/:/iter2/.eldev/docker-home/git/undercover/ silex/emacs:27 sh -c ELDEV_LOCAL=/eldev /eldev/bin/eldev eval 1’

    Output of the /usr/bin/docker run:
    docker: Error response from daemon: create ~/eldev: "~/eldev" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
    See 'docker run --help'.

    /usr/bin/docker run exited with error code 125
    Finished erroneously on Sat Nov 6 19:05:38 2021

Am I doing something wrong?

doublep:
Ah, apparently it's caused by `~` in `$ELDEV_LOCAL` (here it is `~/eldev`). Please fix this too before I merge, because you know the code better.

Also, after working around this I still get yet another error:

    $ ELDEV_LOCAL=/home/paul/eldev eldev -dt docker 27 eval 1
    Started up on Sat Nov 6 19:13:10 2021
    Running on GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 2.24.32)
     of 2020-10-27
    Project directory: ‘/home/paul/eldev/’
    Loading file ‘/home/paul/.eldev/config’...
    Loading file ‘Eldev’...
    Loading file ‘Eldev-local’...
    Executing command ‘docker’...
    Executing ‘eldev-executing-command-hook’...
    Running command ’/usr/bin/docker run --rm -e HOME=/eldev/.eldev/docker-home -u 1000:1000 -v /home/paul/eldev/:/eldev -w /eldev -v /home/paul/.eldev/config:/eldev/.eldev/docker-home/.eldev/config -v /home/paul/.eldev/global-cache:/eldev/.eldev/docker-home/.eldev/global-cache silex/emacs:27 sh -c ELDEV_LOCAL=/eldev /eldev/bin/eldev eval 1’

    Output of the /usr/bin/docker run:
    docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: rootfs_linux.go:76: mounting "/home/paul/.eldev/config" to rootfs at "/eldev/.eldev/docker-home/.eldev/config" caused: mount through procfd: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type.
    time="2021-11-06T19:13:10+01:00" level=error msg="error waiting for container: context canceled"

    /usr/bin/docker run exited with error code 125
    Finished erroneously on Sat Nov 6 19:13:10 2021

LaurenceWarne:
> Ah, apparently it's caused by ~ in $ELDEV_LOCAL (here it is ~/eldev). Please fix this too before I merge, because you know the code better.

Weird, I was using `~` also, maybe my terminal emulator was doing some kind of auto expansion, should be fixed now anyhow. 

> Also, after working around this I still get yet another error:

For some reason I got it into my head that `~/.eldev/config` was a directory :facepalm:, so I think you having an existing file at that location was giving `Are you trying to mount a directory onto a file (or vice-versa)?`.  This should be fixed now also, sorry!

doublep:
Thank you, now it works, so merged! I massaged it a bit before committing, mostly the docstring of the new command.

I will release 0.10 with this new command and some testing support improvements (most important are already committed) next week.

> Thanks for your patience.

Rather thank you. All the time I was afraid you'd say: "Fuck it with all the requests, I'm done here" ;). Then I'm not sure if I would come up with the clever hack for user mismatch and "file created by root" problem caused by it.

LaurenceWarne:
Awesome!  Thanks again for your time and quick responses.

> "Fuck it with all the requests, I'm done here" ;).

This made me laugh :smile:

Issue #55, opened 2021-10-23
GitHub action which tests doublep/eldev on Windows Server 2019 servers has suddenly started to fail connecting to MELPA


ikappaki:
Some `Eldev` tests download packages from `MELPA` (e.g. `buttercup`), but they suddenly started failing to connect to the repostiory. This was only became obvious when an unrelated commit was made to the codebase after a few months of inactivity.

Example error indicates a connection issue to melpa:

```
[03:40.937]  Ran Eldev as `d:/a/eldev/eldev/bin/eldev --setup '(eldev-require-version "999.9")' upgrade-self --dry-run' in directory `d:/a/eldev/eldev/test/project-a/'
[03:40.937]  Stdout contents:
             
[03:40.937]  Stderr contents:
             Process stable.melpa.org not running
             When updating contents of package archive `melpa-stable'
             Failed setup step: evaluating form `(eldev-require-version "999.9")' specified on the command line
             Project `project-a' requires Eldev version 999.9 or newer (this is version 0.9.4snapshot)
             
[03:40.937]  Process exit code: 1
```

Preliminary analysis has shown that this is likely to the Let's encrypt certificate used by https://stable.melpa.org whose [issuer has expired on the 30th of Sep](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/). There appears to be  another certificate that replaces it, but both are visible as described here [in this comment](https://github.com/win-acme/win-acme/issues/1918#issuecomment-911117119).

It appears the Emacs version we install on the GitHub windows 2019 server  comes with `gnutsl-3.6.12`, and can't handle the situation very well:

(output captured by reading the `*Messages*` buffer after `(setq gnutls-log-level 3)`:
```
gnutls.c: [3] ocsp signer: subject `CN=R3,O=Let's Encrypt,C=US', issuer `CN=ISRG Root X1,O=Internet Security Research Group,C=US', serial 0x00912b084acf0c18a753f6d62e25a75f5a, RSA key 2048 bits, signed using RSA-SHA256, activated `2020-09-04 00:00:00 UTC', expires `2025-09-15 16:00:00 UTC', pin-sha256="jQJTbIh0grw0/1TkHSumWb+Fs0Ggogr621gT3PvPKG0="
             
gnutls.c: [3] ASSERT: ../../../gnutls-3.6.12/lib/x509/ocsp.c[gnutls_ocsp_resp_get_single]:1650
             
[...]

PR #56, opened 2021-10-26
Switch to MSYS2 for installing Emacs on MS-Windows in CI (fix for #55)

ikappaki:
Hi,

could you please consider fix for #55 whereby the Emacs MS-Windows precompiled binaries installed with `purcell/setup-emacs@master` GitHub action used in Eldev CI, ships with a GnuTLS version which is susceptible to a bug with expired issuer certificates.

The solution proposed here is to install Emacs as an msys2 package. msys2 itself is already preinstalled on the GitHub Windows 2019 servers, we just utilize it to install the latest Emacs package and set up the global PATH to point to it with an additional GitHub action.

Once a new binary is available in `purcell/setup-emacs`, we can switch back to that installer. The benefit of `purcell/setup-emacs` over msys2 is that you can specify which version to install, while with msys2 you can only download the latest available precompiled version (which can change without notice, i.e. going from 27.2 to 28.5 once that becomes available), i.e. no fine grain control which exact version/binary is being used.

All test pass now on CI.

Thanks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,9 +27,8 @@ jobs:
             emacs_version: '27.2'
           - os: macos-latest
             emacs_version: 'snapshot'
-          # Currently disabled as failing for reasons unrelated to Eldev, see issue #55.
-          # - os: windows-latest
-          #   emacs_version: '27.2'
+          - os: windows-latest
+            emacs_version: '27.2'
 
     steps:
     - name: Set up Emacs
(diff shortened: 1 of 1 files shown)


doublep:
@ikappaki: Hello. Now that Emacs 28.1 is out, can we revert to using `purcell/setup-emacs` also for Windows? Or are there other problems?

ikappaki:
> @ikappaki: Hello. Now that Emacs 28.1 is out, can we revert to using `purcell/setup-emacs` also for Windows? Or are there other problems?

Hi @doublep, yes we can revert, submitted #69 for it. (it is cs090218/setup-emacs-windows we revert to, not purcell/setup-emacs, I had made a mistake in the comments). thanks

PR #60, opened 2021-11-07
Add Documentation for docker Command

LaurenceWarne:
Hi!  Here's what I had locally previously for the `eldev docker` command doc.  No problem if this is something you'd rather put together yourself.

Quick link to rendered README: https://github.com/LaurenceWarne/eldev/tree/docker-doc#using-docker.


doublep:
Thank you. I'll use this as the basis.

Issue #62, opened 2021-12-11
Is it possiable to support XDG Based Directory?


declanqian:
Emacs itself has xdg support (~/.config/emacs/).

doublep:
What do you mean by support here?

declanqian:
Then, you would be surprised if do this search [q=+XDG+support&type=Issues](https://github.com/search?q=+XDG+support&type=Issues) on github.
For The Record, I don't want to push this change though. Feel free to close this issue if you are not into this idea.

In summary, instead of putting a `.eldev` in user Home directory, move eldev related config/data/cache to obey platform-specific directory standards? Cargo (Rust build tool) has a [rfc](https://github.com/rust-lang/rfcs/pull/1615/files) for this.

Also, see a list of tools are supporting from [Archlinux wiki](https://wiki.archlinux.org/title/XDG_Base_Directory#Supported)

doublep:
Currently, Eldev stores configuration mixed up with caches in `~/.eldev`. I.e. there are file `config`, directories `MAJOR.MINOR` per used Emacs version (containing own bootstrapped Eldev package) and, finally, directory `global-cache`, containing packages downloaded from internet archives. What would be the correct setup according to XDG rules?

declanqian:
config:    .config/eldev/config  <-- ~/.eldev/config

For the cache/data
cache: .cache/eldev/
data:  .local/share/eldev/

Maybe borrow some structure from `straight.el`.
Isn't the archives/package source version independent?
Seems to me, only the build output depends on Emacs version

```
tree -L 3 .eldev/
.eldev/
└── 28.0
    └── bootstrap
        ├── archives
        ├── eldev-0.10.3
        └── gnupg
```
```
tree -L 1 straight
straight
├── build-27.2
├── build-27.2-cache.el
├── build-28.0.50
├── build-28.0.50-cache.el
├── build-28.0.60
├── build-28.0.60-cache.el
├── build-29.0.50
├── build-29.0.50-cache.el
├── build-wr
├── build-wr-cache.el
├── links
├── modified
└── repos
```

doublep:
> Isn't the archives/package source version independent?

It is, but `~/.eldev/MAJOR.MINOR` contains byte-compiled installed Eldev package. Those are not version-independent. Just run with `ELDEV_LOCAL` set to an empty value and check directory contents afterwards.

> Maybe borrow some structure from straight.el.

[...]

PR #66, opened 2022-02-13
Include `>' in the list of chars to escape when quoting batch text

ikappaki:
Fixes MS-Windows CI test bootstrap failure due to change in #62.


doublep:
Great, thank you!

PR #71, opened 2022-05-14
Fix/GitHub eldev install win

### LINKED ISSUE #70: webinstall/github-eldev.bat fails to install Eldev properly
Hi,

The `webinstall/github-eldev.bat` fails to work as a GitHub workflow, although the `eldev.bat` is successfully installed, it fails to publish the location it is installed at.

To reproduce, create a github workflow file, that installs and runs `eldev.bat` on MS-Windows:

```yml
name: CI

on: [push, pull_request]

jobs:
  test:
    runs-on: ${{matrix.os}}

    strategy:
      matrix:
        os: [windows-latest]
        emacs_version: [ '28.1']

    steps:
    - name: Set up Emacs on Windows
      uses: jcs090218/setup-emacs-windows@master
      with:
        version: ${{matrix.emacs_version}}

    - name: Install Eldev on MS-Windows
      run: |
        curl.exe -fsSL https://raw.github.com/doublep/eldev/master/webinstall/github-eldev.bat | cmd /Q
 
    - name: Check out the source code
      uses: actions/checkout@v2

    - name: Test the action worked.
      run: |
        eldev help
``` 

It fails with:

```
...

2s
Run curl.exe -fsSL https://raw.github.com/doublep/eldev/master/webinstall/github-eldev.bat | cmd /Q
Microsoft Windows [Version 10.0.20348.707]
(c) Microsoft Corporation. All rights reserved.

D:\a\eldev\eldev>
10s
2m 7s
Run eldev help
eldev: D:\a\_temp\66bfe8f7-dc7c-4248-87dc-f22aaa1afcc2.ps1:2
Line |
   2 |  eldev help
     |  ~~~~~
     | The term 'eldev' is not recognized as a name of a cmdlet, function, script file, or executable
     | program. Check the spelling of the name, or if a path was included, verify that the path is correct
     | and try again.

Error: Process completed with exit code 1.
...
```

Although the tool is installed, it is not in the `PATH`. Not sure if this has ever worked before. PR to follow.

ikappaki:
Hi,

can you please consider patchfor github-eldev.bat installations script that fixes #70.

It appears that the MS-Windows github installation script does not set the eldev installation path correctly, it adds additional space at the end when running the `echo`, and thus the end path has an additional space at the end (!), which is invisible to the naked eye .... Not sure how this ever worked.

I've also included a new CI workflow that tests the github installation scripts across the supported architectures. This is a separate file, so that it does not interfere with the actual eldev tests.

Thanks,

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,20 +122,29 @@ jobs:
           await io.rmRF(expandTilde("~/.local/bin"));
 
     # GitHub-specific installation specifically for workflows.
-    - name: Test GitHub-specific installation
-      uses: actions/github-script@v3
-      with:
-        script: |
-          const fs = require('fs');
-          const exec = require('util').promisify(require('child_process').exec);
-          const {getRawUrl,downloadScript,expandTilde} = require(`${process.env.GITHUB_WORKSPACE}/.github/workflows/util.js`);
-          if (fs.existsSync(expandTilde("~/.local/bin/eldev")))
-              throw("~/.local/bin/eldev already exists");
-          await io.mkdirP('tmp');
-          const script = await downloadScript(github, context, "webinstall/github-eldev", 'tmp');
-          const eldev_bin_url = getRawUrl(context,"bin/eldev");
-          await exec(`${script} ${eldev_bin_url}`);
-          const eldev_out = await exec(expandTilde("~/.local/bin/eldev"));
-          console.log(eldev_out.stdout);
-          await io.rmRF(expandTilde("~/.local/bin"));
+    - name: Test GitHub-specific installation (1/2)
+      if: "!startsWith (matrix.os, 'windows')"
+      run: |
+        if command -v eldev &> /dev/null
+        then
+         echo eldev is already installed.
+         exit 1
+        else
+          curl -fsSL https://raw.github.com/doublep/eldev/master/webinstall/github-eldev | sh
+        fi
 
+    - name: Test GitHub-specific installation on MS-Windows (1/2)
+      if: startsWith (matrix.os, 'windows')
+      run: |
+        if (Get-Command "eldev" -ErrorAction SilentlyContinue)
+        {
+          throw 'eldev is already installed.'
+        }
+        else
+        {
+          curl.exe -fsSL https://raw.github.com/doublep/eldev/master/webinstall/github-eldev.bat | cmd /Q
+        }
+
+    - name: Test GitHub-specific installation (2/2)
+      run: |
+        eldev help && rm ~/.local/bin/eldev*
--- a/webinstall/github-eldev.bat
+++ b/webinstall/github-eldev.bat
@@ -23,5 +23,5 @@ mkdir %ELDEV_BIN_DIR%
 
 curl.exe  -fsSL %URL% -o %ELDEV_BIN_DIR%\eldev.bat || exit /b
 
-echo %ELDEV_BIN_DIR% >> %GITHUB_PATH%
+echo %ELDEV_BIN_DIR%>> %GITHUB_PATH%
 


ikappaki:
It appears that having a separate github-eldev install test workflow clatters the CI view too much. I will try to merge it in test.yml.

ikappaki:
So, I've merged the new github-eldev install tests into `test.yml`, replacing the existing github-eldev installation test which only tested that `eldev.bat` was installed, but not if it can be executed.

The window test is currently fails on CI because the actual fix it tests is not in master yet, where the `github-eldev.bat` is sourced from.

(I will squash when you give the go ahead, I wanted to keep the narration above valid)

doublep:
> Not sure how this ever worked.

Possibly it never did, I'm not sure if anyone uses this script on Windows. Does e.g. CIDER?

> I've merged the new github-eldev install tests into test.yml

Perfect, just wanted to ask.

> The window test is currently fails on CI because the actual fix it tests is not in master yet, where the github-eldev.bat is sourced from.
> \-      uses: actions/github-script@v3

I believe this was specifically converted to JS because of Windows support, see commit `44c49ae Use javascript-action instead of platform dependent Shell commands`. Are you sure it is not needed now? Can it be that it fails because of that?

@juergenhoetzel: do you have any comments on this PR? It heavily touches code that you wrote.

ikappaki:
> > The window test is currently fails on CI because the actual fix it tests is not in master yet, where the github-eldev.bat is sourced from.
> > -      uses: actions/github-script@v3
> 
> I believe this was specifically converted to JS because of Windows support, see commit `44c49ae Use javascript-action instead of platform dependent Shell commands`. Are you sure it is not needed now? Can it be that it fails because of that?

Indeed, it is different, but the new test is more conclusive I think:
1. Existing test: run webinstall script and check if elder scripts has been installed at the expected location.
2. New test: run web install script and test that you can call the installed eldev script successfully.

The problem with 1 was that although it checks whether the elder scripts was installed at the specified location, it does not really test if the script can be called and executed.

2 is better in the sense that it is more complete, that it also tests that the user can call the script successfully after installation. Do we also care to check if the script is installed at the local/bin location? in which case I think we can also keep the javascript  code (slightly modified) as step 1.5 of 2.

Thanks

doublep:
> Indeed, it is different, but the new test is more conclusive I think:
>
> 1. Existing test: run webinstall script and check if elder scripts has been installed at the expected location.
> 2. New test: run web install script and test that you can call the installed eldev script successfully.
>
> [...]
>
> 2 is better in the sense that it is more complete, that it also tests that the user can call the script successfully after installation.

Agreed, this is better. I didn't look before because I usually just leave Windows improvements to you and Jürgen, since I don't even have access to that OS. But now I have checked the changes closer.

> Do we also care to check if the script is installed at the local/bin location?

Not really important. On the other hand, since we remove installed Eldev to make other installation tests possible, we need to know what to remove. By the way, doesn't `rm` itself work as a test? I.e. won't it fail itself (and thus the workflow with it) if there is nothing to remove? Also, would be nice to test that Eldev cannot be executed *before* having run `github-eldev[.bat]`.

If you can get rid of JS here, would it also be possible for other installation method tests? I see there are three in `test.yml`.

ikappaki:
> > Do we also care to check if the script is installed at the local/bin location?
> 
> Not really important. On the other hand, since we remove installed Eldev to make other installation tests possible, we need to know what to remove. By the way, doesn't `rm` itself work as a test? I.e. won't it fail itself (and thus the workflow with it) if there is nothing to remove? Also, would be nice to test that Eldev cannot be executed _before_ having run `github-eldev[.bat]`.

Yes, `rm` is a test by itself, it will fail if `eldev*` is not there. Also I realised that both `eldev help` and `rm` should be on the same line, since we need both to succeed.  I have also updated the github eldev installation jobs as suggested, to test for the presence of existing `eldev `installation and error out if they find one.

> 
> If you can get rid of JS here, would it also be possible for other installation method tests? I see there are three in `test.yml`.

Most probably yes, but I won't have time to look at the other js job with this PR unfortunately.

ikappaki:
> > Also I realised that both eldev help and rm should be on the same line, since we need both to succeed.
> 
> Don't GitHub workflows fail if any command in a script fails, not only the last one?
> 

I don't think this is the always case; github invokes the shell of choice on the script and just inspects its return value. It's up to the shell to check each individual command and fail if is design so.

> > Most probably yes, but I won't have time to look at the other js job with this PR unfortunately.
> 
> No problem. Would not quite be related to this PR anyway.
> 
> Please squash the commits.

Done, thanks!

doublep:
Thank you. Let's hope that the failure is really gone once merged.

Issue #72, opened 2022-05-15
Eldev fails to bootstrap on Emacs 26


ikappaki:
It throws a `Trailing garbage following expression` error.

1. Download and locally extract Emacs 26 from http://ftp.gnu.org/gnu/emacs/windows/emacs-26/.
2. Open a Command prompt
3.  cd to eldev src
4. Point to 26.3 release and invoke eldev:
```batch
D:\src\eldev>set ELDEV_LOCAL=.

D:\src\eldev>set ELDEV_EMACS=d:/emacs-26.3-x86_64/bin/emacs.exe

D:\src\eldev>bin\eldev.bat
Trailing garbage following expression:
```
Although Emacs 26 is an old version, I believe it is good to have support for those Eldev users that have CI workflows that support older versions.

PR to follow.

doublep:
Should be fixed by PR #73, right?

ikappaki:
> Should be fixed by PR #73, right?

Correct, as implicitly referenced in this issue.

PR #73, opened 2022-05-15
Fix Emacs 26 bootstrap quoting issue with trailing newline

ikappaki:
Fixes Emacs 26 issue with the command line `--execute "..."` option that does not allow newline characters at the end of the quoted command (#72).

Currently Eldev always appends a newline when `eldev-bat-quote`ing the bootstrap code (where it shouldn't add anything more to it).  The patch remove this extraneous newline character. It also ensures that it removes any newlines from the bootstrap code in `eldev.bat.in` by trimming the bootstrap contents.

Finally, there is a unexplained bizarreness with the .bat file reader that it will not allow the second `--execute` to be on its own line if `"` is not the first line character in our case, and thus all command line options after the first `--execute` have been merged into one line. 

I have not included a test since this will require running Emacs 26 Windows in GitHub CI, which would be too much to ask for. I would say older Emacs versions on Windows are on a best effort support, and if it silently breaks, so be it until someone complains.

Thanks

### DIFF
--- a/bin/eldev.bat
+++ b/bin/eldev.bat
@@ -99,10 +99,7 @@ REM the newline variable above MUST be followed by two empty lines.
         (package-activate-1 eldev-pkg) !NL!^
         (advice-remove #'load autoloads-disabler)))) !NL!^
   (require 'eldev) !NL!^
-  (eldev-start-up)) !NL!^
-"  ^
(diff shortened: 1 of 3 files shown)


doublep:
Please just squash the commits and I will merge this.

ikappaki:
> Please just squash the commits and I will merge this.

Done, thanks.

doublep:
Thank you.

PR #75, opened 2022-08-16
Fix nongnu unstable repo url

xFA25E:


doublep:
Thank you for spotting this.

Issue #50, opened 2021-09-30
Packaging with "tar.exe" on Windows


dominiksta:
Hi,

I am trying out Eldev for the first time and I am really liking it so far. However, I think I may have stumbled upon a bug when trying to create a tarball on windows - meaning I cannot build a package.

Here is the output of `eldev -vt package`:
```
Started up on Thu Sep 30 16:06:59 2021
Running on GNU Emacs 27.2 (build 1, x86_64-w64-mingw32)
 of 2021-03-26
Project directory: `c:/Users/dominik/.emacs.d/straight/repos/mvtn.el/'
No file `c:/Users/dominik/.eldev/config', not applying user-specific configuration
Loading file `Eldev'...
Using package archive `melpa-stable' at `https://stable.melpa.org/packages/' with priority 200
Using package archive `melpa-unstable' at `https://melpa.org/packages/' with priority 100
Reading target dependencies from file `.eldev/27.2/target-dependencies.build'...
Saving target dependencies to file `.eldev/27.2/target-dependencies.build'...
No file `Eldev-local', not customizing build
Executing command `package'...
Contents of package archive `melpa-stable' has been fetched already
Contents of package archive `melpa-unstable' has been fetched already
Not fetching contents of other archive(s) as redundant
All project dependencies (including those for set `build') have been installed already or are local
Loading file `mvtn-autoloads.el'
Generating build target list for standard fileset `main'
Ignored potential target: `mvtn-pkg.elc' (1)
Reading target dependencies from file `.eldev/27.2/target-dependencies.build'...
Building plan: `dist/mvtn-0.1.tar', `:package'
PACK     -> dist/mvtn-0.1.tar
Packaging the following files: `mvtn-ag.el', `mvtn-compat.el', `mvtn-file-helpers.el', `mvtn-link-buttons.el', `mvtn-pkg.el', `mvtn-rg.el', `mvtn-tag-addons.el', `mvtn-templates.el', `mvtn.el' (9)
Full command line to create package tarball:
[...]

PR #80, opened 2022-12-30
eldev doctor check for tar on MS-Windows

ikappaki:
Hi,

could you please consider patch to add a new doctor check for the compatibility of `tar` on MS-Windows. It addresses #50.

It returns the following warn when it can't use tar

```
Is the tar executable found compatible with Eldev on MS-Windows? NO

The tar executable found at `c:/Program Files/Git/usr/bin/tar.exe` is not
compatible with Eldev, please use the one provided with MS-Windows (typically
located at `c:\Windows\system32\tar.exe`), by either rearranging the entries
in the PATH environment variable to pick that up first, or by setting the
`eldev-tar-executable' in `Eldev-local` to something like

(setf eldev-tar-executable "C:/Windows/system32/tar.exe").

Ran 1 doctest, it generated a warning
```

I have also added a test.

Thanks


doublep:
> The tar executable found at `c:/Program Files/Git/usr/bin/tar.exe` is not compatible with Eldev

This largely reads as "Eldev sucks". I would rephrase this to mention what's wrong about it, i.e. mishandling of drive letters in absolute Windows paths.

More importantly, is there no way to convince GNU tar to behave sanely? E.g. an option, an environment variable, some sort of escaping in paths? Eldev already has a shitton of workarounds everywhere, why not add another one? Alternatively, maybe we could use a non-absolute path for tarring on Windows only if possible (i.e. if on the same drive).

ikappaki:
> > The tar executable found at `c:/Program Files/Git/usr/bin/tar.exe` is not compatible with Eldev
> 
> This largely reads as "Eldev sucks". I would rephrase this to mention what's wrong about it, i.e. mishandling of drive letters in absolute Windows paths.

Sure, this is why code reviews are so important :)

> More importantly, is there no way to convince GNU tar to behave sanely? E.g. an option, an environment variable, some sort of escaping in paths? Eldev already has a shitton of workarounds everywhere, why not add another one? Alternatively, maybe we could use a non-absolute path for tarring on Windows only if possible (i.e. if on the same drive).

There's this `--force-local` flag for GNU tar that appears to do the trick. 

Let me try to summarise the problem to put any prospective solution into perspective.

Problem: It is common for MS-Windows program to have a (graphical) installer that installs programs into a directory that the user should not be concerned about. Usually, this directory's bin folder is placed at the start of the `PATH` environment variable (without even the user realizing), so that he program can be invoked by either clicking the program's icon or typing the program names on the command line.  Some programs, such as `git for windows`, are primarily developed on Linux and only ported on MS-Window afterwards. They may have additional binary tools dependencies that usually bundled with, and installed alongside the program. These binary dependencies may not be 100% compatible with MS-Windows but they are most likely to work flawlessly for at least the range of operations required by the main program. In our example case, `git for windows` comes with a ported version of GNU `tar` that does not work out for the box on MS-Windows and while it is placed at the front of the `PATH`, courtesy of the `git for windows` installation program,  is causing `Eldev` to pick it up in favor of the standard battle tested `tar` which comes with MS-Windows.

Possible options to workaround both the general and specific GNU tar issue

1. [General solution] Prefer the battle tested `tar` program that comes with MS-Windows over any other ported `tar` version that was installed as a side effect of some other unrelated program's installation. This can be perhaps done by having `eldev-tar-executable` fn find `$WINDIR/system32/tar.exe` unless `eldev-tar-executable` is set explicitly by the user. The previous discussion on this solution suggested to use the absolute `c:\Windows\System32\tar.exe` path which you advised is a terrible idea (perhaps using the `WINDIR` env variable makes it  more acceptable?).
2. [General solution] Have `eldev doctor` check whether the current tar executable found first in PATH is working on MS-Windows, if not, advise user to rejiggle the PATH order so that the windows `tar` program comes first.
3. [Gnu tar specific solution] A naive implementation would be to have the `eldev-tar-executable` fn to check whether the first tar found in PATH is GNU tar (perhaps by executing `tar --version`), and, if so, set a new var flag to indicate the same . We then update every reference of this fn or `eldev-tar-executable` variable to also check about whether the new flag has been set, and if so, invoke tar with the `--force-local` option. It would be requiring some effort to write a test for this, but should be achievable.  

What do you think?

Thanks

doublep:
Thanks for the detailed analysis, now I understand the issue much better. However, I still don't feel like overriding what `PATH` says and (semi-)hardcoding which `tar` is to be used.

I suggest going with 2 + 3. Implement whatever hacks are necessary to make Eldev work even with GNU tar (3). But still add a check to the doctor (2) that would complain about GNU tar being in path: even if Eldev itself is able to cope with it now, it would potentially be problematic for general use on Windows.

Also:

> It would be requiring some effort to write a test for this, but should be achievable.

Tests are nice to have, but not an absolute must. In this case a test can be skipped as long as you say that it works fine on your machine.

PR #108, opened 2024-11-27
Fix missing `setf` symbol on <24.3

KeyWeeUsr:
```sh
docker run --rm -u 1000:1000 docker.io/silex/emacs:24.1 sh -c 'emacs --batch --eval "(setf)"'
```

vs

```sh
docker run --rm -u 1000:1000 docker.io/silex/emacs:24.1 sh -c "emacs --batch --eval \"(progn (require 'cl) (setf (buffer-name) (buffer-name)))\""
```

and `cl-lib` doesn't exist there yet I presume, so it might trigger a deprecated `cl` message somewhere if `when` isn't adhered to, so I miiight be missing some flag or something here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        emacs_version: ['24.4', '24.5',
+        emacs_version: ['24.1', '24.4', '24.5',
                         '25.1', '25.2', '25.3',
                         '26.1', '26.2', '26.3',
                         '27.1', '27.2',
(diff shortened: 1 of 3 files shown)


doublep:
What is the goal here? Eldev supports Emacs from 24.4 (it's documented), there are _many_ more things that will fail on yet older versions, `setf` being the smallest problem. Besides, Emacs 24 is so old I doubt anyone anywhere uses it.

KeyWeeUsr:
Ah, my bad. The PR was for fixing it to run on 24.1, however now that you mention the docs, it's probably the other way around and these should be fixed in code :thinking:

PR #109, opened 2024-11-27
Bump required Emacs versions to 24.4

KeyWeeUsr:
Match the required versions to the documented one.

Closes #108


doublep:
Thank you. When I started this project I planned to support Emacs 24 and up, but then found out that several things I need only work starting with 24.4. Simply forgot to update the requirements in some places, though documentation and package declaration were correctly mentioning 24.4.

